PR #2, opened 2018-10-14
Some proposed changes

ZacLN:
Notably:

- Uses struct's for symbol information rather than structs throughout, this is faster for StaticLint
- Process for loading a package is: 1) msg to server to load; 2) server loads and saves to disc sending back  the paths of `jstore` files; 3) client side loads from disk
- Some redirecting of stdout [some packages were printing output on loading]
- two lists of packages available 1) installed packages; 2) install + un-loadable dependencies [this means when we open an installed package we have access to the imported dependencies]
- Always load Base/Core into a const global so we have access to those without any store
- StaticLint now includes this as a dependency (https://github.com/ZacLN/StaticLint.jl/pull/11) [This is necessary, StaticLint has no functionality without SymbolServer]
- `load_module` (copied across from StaticLint) marks references to dependencies in their symbol tables and loads them

### DIFF
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -3,7 +3,9 @@ module SymbolServer
 export SymbolServerProcess
 export getstore
 
-using Serialization
+using Serialization, Pkg
+
+include("clientprocess/from_static_lint.jl")
 
 mutable struct SymbolServerProcess
     process::Base.Process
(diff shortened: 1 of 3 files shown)


ZacLN:
I have things mostly working outside of a vscode setting (https://github.com/ZacLN/StaticLint.jl/tree/symserver, https://github.com/ZacLN/LanguageServer.jl/tree/symserv) but am running into trouble with serialisation between the SymbolServer and language server when run from the the vscode package directory:

```julia
ERROR: LoadError: EOFError: read end of file
Stacktrace:
 [1] read at ./iobuffer.jl:192 [inlined]
 [2] read at ./stream.jl:758 [inlined]
 [3] deserialize at /home/zac/julia/usr/share/julia/stdlib/v1.1/Serialization/src/Serialization.jl:703 [inlined]
 [4] deserialize(::Base.PipeEndpoint) at /home/zac/julia/usr/share/julia/stdlib/v1.1/Serialization/src/Serialization.jl:700
 [5] top-level scope at /home/zac/.vscode/extensions/julia-vscode/scripts/languageserver/packages/SymbolServer/src/clientprocess/clientprocess_main.jl:8
etc...

ERROR: LoadError: KeyError: key SymbolServer [4cc854a7-cf14-540f-aa0e-47853c5fab2f] not found
Stacktrace:
 [1] getindex at ./dict.jl:478 [inlined]
 [2] root_module at ./loading.jl:895 [inlined]
 [3] deserialize_module(::Serialization.Serializer{IOStream}) at /home/zac/julia/usr/share/julia/stdlib/v1.1/Serialization/src/Serialization.jl:831
etc...


```
Any ideas? I think there's somehow a mismatch between the versions of SymbolServer on the client and server side- vscode setup is at: https://github.com/JuliaEditorSupport/julia-vscode/tree/symserv

PR #3, opened 2018-10-15
WIP Handle environments better

davidanthoff:
@ZacLN This is a bit WIP, but I wanted to make sure you see what I'm playing with :)

The general idea is this: if we edit a folder that is a package *and* that folder is also deved in the active environment, then we actually want to handle things in the same way that base handles tests and builds: we want to load things in a new temporary environment, in which all the dependencies that the project has become top level dependencies, but at the same time use the versions that are in the manifest in the environment. That is what all this mess here is about...

But I think essentially that will allow us to resolve the proper names: Say I have an environment that contains a deved version of ``Foo``, and ``master`` of ``Bar``. Lets also assume ``Foo`` has ``Bar`` and ``Bar2`` in its ``REQUIRE`` file. If we now open ``Foo`` in VS Code, then I think we need the following things to work: a) we need to be able to ``import Bar2`` from our symbol server. That in itself won't work, because ``Bar2`` is not a top level dep in the environment. So for the symbol server run, we actually need to lift that to the top level. b) we also need to lift ``Bar`` to the top level, but in this case it already is, and we need to make sure that we actually use the version that is part of the environment. All of this is exactly the same set of issues that base handles for testing and building, and this PR here just mimics what they do.

I still need to clean things up a bit, and then integrate it properly into LS... Will need at least another evening for that, I think.

Some other smaller random things:
- I think we should not have ``using SymbolServer`` run in the symbol server process, because we can't be sure that this package will be available. So instead I'm just using ``include``, that should always work.

### DIFF
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -10,24 +10,22 @@ include("clientprocess/from_static_lint.jl")
 mutable struct SymbolServerProcess
     process::Base.Process
 
-    function SymbolServerProcess(;environment=nothing, depot=nothing)
+    function SymbolServerProcess(;environment=nothing, depot=nothing, project=nothing)
         jl_cmd = joinpath(Sys.BINDIR, Base.julia_exename())
         client_process_script = joinpath(@__DIR__, "clientprocess", "clientprocess_main.jl")
 
-        env_to_use = copy(ENV)
-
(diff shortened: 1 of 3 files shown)


davidanthoff:
Outdated.

PR #10, opened 2018-11-20
Store caches of base/core - take 2

davidanthoff:
Does the same thing as #9, but aligns the code more with the existing pattern.


davidanthoff:
I'm just going to merge this, given that it does the same thing as the one @ZacLN proposed.

ZacLN:
Yep looks good

PR #11, opened 2018-12-25
make response handling safer (?)

ZacLN:
check response is a tuple before accessing it- doesn't address why we'd not be getting a tuple but will error correctly


ZacLN:
Yep it would

PR #19, opened 2019-03-04
Improve stdout redirection in client process

Ashymad:
While researching https://github.com/JuliaEditorSupport/LanguageServer.jl/issues/330 I've found that the server waits for response from the client process indefinitely. This is because if there is an error while loading package in the :load_package case, the stdout is still redirected while the catch block executes, which causes the client process to print nothing and, in turn, for the  master to wait for response indefinitely. This PR fixes this and makes the code little bit nicer (IMO).

### DIFF
--- a/src/clientprocess/clientprocess_main.jl
+++ b/src/clientprocess/clientprocess_main.jl
@@ -9,6 +9,14 @@ const depot = Dict("manifest" => c.env.manifest,
                     "installed" => (VERSION < v"1.1.0-DEV.857" ? c.env.project["deps"] : c.env.project.deps),
                     "packages" => Dict{String,Any}())
 
+if Sys.isunix()
+    global const nullfile = "/dev/null"
+elseif Sys.iswindows()
+    global const nullfile = "nul"
+else
+    error("Platform not supported")
+end
+
 while true
     message, payload = deserialize(stdin)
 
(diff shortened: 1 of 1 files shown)


src/clientprocess/clientprocess_main.jl
```diff
@@ -34,15 +34,12 @@ while true
             end
             serialize(stdout, (:success, pkgs))
         elseif message == :load_package
-            ostdout = stdout
-            (outRead, outWrite) = redirect_stdout() # seems necessary incase packages print on startup
-            SymbolServer.import_package(payload, depot)
+            redirect_stdout(open("/dev/null", "w")) do # seems necessary incase packages print on startup
```
davidanthoff: `/dev/null` doesn't work on Windows...

Maybe just sticking with the old code but adding some `try finally` would also do the trick?

Ashymad: Oh yeah, forgot about that... I see a few options:
 1. We could check the platform and use `"nul"` special file on Windows
 2. We could add inner try/catch/finally block in the `:load_package` case, rethrow the error in `catch` and redirect `stdout` in `finally`
 3. We could simply redirect `stdout` to `ostdout` in the catch block regardless if it was redirected earlier or not (don't know if not closing the pipe ends would be a problem)

davidanthoff: I like the `nul` option, that seems cleanest!

Ashymad:
Finally I've found some time to work on this. Now it should work on Windows as well

PR #22, opened 2019-03-28
Restore compatibility with Julia 1.1

tribut:
This commit resolves the following issues:

 * UUIDs were used as keys for a `Dict{String,Any}`, leading to an exception being thrown.
 * `depot["manifest"]` is now indexed by UUID instead of by name.
 * `nullfile` was never closed (not strictly an issue with Julia 1.1, but the fact that output redirection was not restored lead to the silent failure / freeze with Julia 1.1)

Fixes https://github.com/JuliaEditorSupport/julia-vscode/issues/647

### DIFF
--- a/src/clientprocess/from_static_lint.jl
+++ b/src/clientprocess/from_static_lint.jl
@@ -31,7 +31,7 @@ end
 
 struct structStore <: SymStore
     params::Vector{String}
-    fields::Vector{String}    
+    fields::Vector{String}
     ts::Vector{String}
     methods::Vector{MethodStore}
     doc::String
@@ -85,7 +85,7 @@ function load_module(m, pkg, depot, out)
                         else
                             out.vals[first(dep)] = first(dep)
                         end
-                        # the above make reference to the name of the module, may have to change to uuid 
+                        # the above make reference to the name of the module, may have to change to uuid
                     catch err
                     end
                 end
@@ -117,7 +117,6 @@ function load_module(m, pkg, depot, out)
                 end
             elseif x isa Module && x != m # include reference to current module
                 if parentmodule(x) == m # load non-imported submodules
-                    
                     out.vals[String(n)] = ModuleStore(String(n))
                     load_module(x, pkg, depot, out.vals[String(n)])
                 end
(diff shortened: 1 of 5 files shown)


src/clientprocess/from_static_lint.jl
```diff
@@ -162,3 +162,11 @@ function save_store_to_disc(store, file)
...
     close(io)
 end
+
+function pkg_name(pkg)
+    return first(pkg)
+end
+
+function pkg_uuid(pkg)
+    return string(last(pkg))
+end
```
davidanthoff: I think it is just the other way around, i.e. isn't the first element in the `Pair` the uuid and the last one the name?

src/clientprocess/from_static_lint.jl
```diff
@@ -72,18 +72,18 @@ end
 function load_module(m, pkg, depot, out)
     out.doc = string(Docs.doc(m))
     out.exported = Set{String}(string.(names(m)))
-    if haskey(depot["manifest"], first(pkg))
-        for pkg1 in depot["manifest"][first(pkg)]
-            if pkg1["uuid"] == last(pkg)
+    if haskey(depot["manifest"], pkg_name(pkg))
```
davidanthoff: I'm 90% sure that the dict in `depot["manifest"]` is uuid->name.

So I think we also need to make sure that when we import that dict, that we convert the uuids on julia 1.1. to strings in that dict, otherwise the key lookup won't work. I think that probably needs to be done [here](https://github.com/JuliaEditorSupport/SymbolServer.jl/blob/master/src/clientprocess/clientprocess_main.jl#L8).

I think that was the general problem when I looked last: we have a bunch of dicts where the key type is string on julia 1.0 but `UUID` on julia 1.1, because we just get those dicts directly from Pkg. I think we need to convert all those dicts to string->whatever for it to work with both julia 1.0 and 1.1

tribut:
As discussed on slack, in addition to UUIDs no longer being Strings, the way the manifest was indexed  also changed (old: names, new: uuid). This also explains the confusion in the first version of this PR.

davidanthoff:
Alright, fantastic! I reviewed it and it looks good, and I also think we should just cut a beta/alpha release of the extension and ask folks to try this out in the wild.

polypus74:
Any progress on this by any chance?

davidanthoff:
Yes, this shipped.

PR #23, opened 2019-04-02
Checkdirty

ZacLN:
This makes quite a few (hopefully simplifying..) changes. 
- SymbolServerProcess now holds a Pkg.Types.Context instance from the server listing available packages as well as the loaded package depot. Storing the loaded packages inside this instance is intended to allow us update them if changes are detected to disc files (i.e. environment manifest file or, for deved packages, the source files).
- I've stripped out some unneeded methods for checking available packages.
- A slightly safer approach to loading .jstore files.
- Packages are loaded using a Pkg.Type.PackageEntry rather than name/uuid combo.
- ModuleStore now includes `ver` and `sha` field. These are to be used for checking whether a cache is stale.
- `get_dir_sha` is an (extremely bad) attempt to get a unique identifier for a deved packages source files.

Note: this needs to be checked for backward compatability with respect to Pkg's internals.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
-julia 1.1
+julia 1.0
 SHA
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 1.0
 SHA
+LibGit2
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
-julia 0.7
+julia 1.1
+SHA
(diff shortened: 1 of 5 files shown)


davidanthoff:
I wonder whether we need to first add some tests to this package? Testing that things work on both julia 1.0 and 1.1 gets very, very cumbersome otherwise, at least it was for getting all of this to work with julia 1.1. What do you think?

I'll take a proper look at this PR later!

PR #36, opened 2019-08-11
Refactor for parallel use

ZacLN:
This refactors (and simplifies) the existing code. Instead of launching a distinct process this PR provides for the package to be loaded in a worker process. This allows for the parallel caching of packages as follows:

```julia
using Distributed

@everywhere using SymbolServer

server = SymbolServerProcess()
missing_pkgs = SymbolServer.disc_load_project(server) # load available caches
pkg_uuids = collect(keys(missing_pkgs)) # list of package UUIDs that need to be reloaded
reload_request = @spawnat worker_proc_id SymbolServer.cache_packages_and_save(server.context, pkg_uuids)

while true # Some loop 
    if isready(reload_request) # check reloading is done
        for uuid in fetch(reload_request)
            SymbolServer.disc_load(server.context, uuid, server.depot)
         end
    end
end
```

@davidanthoff , thoughts?


davidanthoff:
I might have missed it, but does this preserve the env handling that we just added? I don't see how it would, to be honest.

Also, I think this would mean that the package loading happens in the LS depot, right? I think we should try to avoid that, having the precompile caches for those two things separated I think is a major benefit in terms of stability.

Finally, I think loading in parallel will probably mean we run into race conditions with precompile, right? So I'm not really sure that is in general a good idea? Not sure about that, though.

ZacLN:
With this env handling is done by simply changing the environment on the worker process 
```
@fetchfrom workerid Pkg.activatge(env_location)
env_context = @fetchfrom workerid Pkg.Types.Context()
```

I'm not clear what you mean by in the 'LS depot', do you mean the main process running the LS? If so then no. 

To make things (hopefully!) a bit clearer the process for use with LanguageServer is as follows:
1. LS initiates
2. LS sends the environment path to the worker.
3. WORKER activates environment.
4. WORKER returns `Context` to LS (listing available packages).
5. LS loads caches for these packages from the disc store.
6. LS sends to WORKER a list of packages that were not cached or are out of date.
7. LS begins main loop and continues with initialisation process.
8. Worker attempts to cache packages from 6 and returns a list of newly cached packages.
9. LS main loop notices the WORKER task is ready and loads new caches from disc

There's no race condition with precompilation

davidanthoff:
Ah, ok, so there is only one WORKER and it loads packages.

So is the main difference between this and the old implementation that the LS doesn't wait for the symbol server to finish loading packages? That seems super useful, but couldn't we do that with the old design as well?

Re depots: The current design is that the LS process uses a depot distinct from the default julia depot ([set here](https://github.com/julia-vscode/julia-vscode/blob/f5847b936e8d6b6aadd2427b92c250aac0db7860/src/extension.ts#L154)), but the symbol server process uses a different depot, namely the default julia depot ([set here](https://github.com/julia-vscode/SymbolServer.jl/blob/54e5263f30720a4629d7753f8b2ef9034a5c49ac/src/SymbolServer.jl#L33)). I think that has greatly helped with stability, and I don't think that is the case with this PR, right?

ZacLN:
We could do this with the old code but it requires using/writing/maintaining our own  implementation for running a parallel task when there already exists a facility within the stdlib. 

I haven't illustrated it in the example code above but we should be be able to explicitly set the depot path on the worker as you normally would in a serial process. It's same as how setting th environment works with this PR (i.e. `@spawn workerid Pkg.activate(location)`) and is kind of the central point - try to make this package as simple as possible and use the existing interfaces for the operations they're intended for.

ZacLN:
Any further thoughts on this @davidanthoff? https://github.com/julia-vscode/LanguageServer.jl/pull/381/commits/3abcba7cf858f2704ea4566dfdb201ce9d2450d1 shows how the depot path is set on the LanguageServer side (requires an [edit](https://github.com/julia-vscode/julia-vscode/commit/905de46ff36c103dc02f31e448d2d52214487323) to the `julia-vscode` side of things to ensure we're passing  the server the depot directory).

ZacLN:
-> #55

PR #45, opened 2019-10-18
Handle lazy initialized DataType.types field properly

kwdye:
Commit a8c8377cc43f024b0c02200f1cd7e0c1ab83caca to the main julia repo made the DataType.types field lazily initialized, so accessing it directly is not correct.  Base.datatype_fieldtypes() will populate and get this field properly.  Simply checking for whether :types is defined is not sufficient.

### DIFF
--- a/src/symbols.jl
+++ b/src/symbols.jl
@@ -154,10 +154,10 @@ function get_module(c::Pkg.Types.Context, m::Module)
             elseif t isa DataType
                 if t.abstract || t.isbitstype
                     out.vals[String(n)] = DataTypeStore(string.(p), [], [], [], _getdoc(x))
-                elseif isdefined(t, :types) && !isempty(t.types) && !Base.isvatuple(t)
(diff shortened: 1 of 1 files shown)


davidanthoff:
Could merge `master` into this branch and then we can see whether tests pass?

Also, I think `datatype_fieldtypes` was added sometime after Julia 1.0.x, right? So we'll probably need a `VERSION` check here so that this code works on all versions of Julia, starting with Julia 1.0.4.

Oh, and if there is some chance that you could add a test for this particular case, it would be awesome! In particular with these things that differ between julia versions, it is otherwise super easy to not catch problems. But if too complicated, no problem. Given how dismal our test coverage is, we might as well add one more thing that is not properly covered by tests ;)

ZacLN:
Thanks, the content of this was brought in by https://github.com/julia-vscode/SymbolServer.jl/commit/7916f584e319554532995811cb6f610cabba88e6#diff-de08f51fe93b9406046593271d4e89c3R239

PR #55, opened 2019-11-29
refactor for parallel use

ZacLN:
Supercedes #36 (rebased on master).


davidanthoff:
I have to admit that I'm still not convinced that we should use the Julia distributed infrastructure for this. I'm worried that we'll run into conflicts where for example packages behave differently if they detect that they run on a child node (I know I've done that in some packages).

I also think that we finally got the whole communication story to be bug free, and so I'm a bit wary to rewrite everything again :)

Having said that, I'm _very_ much in favor of async/parallel. I think that can be achieved super easily with the existing model, though, without any need to use the distributed infrastructure. I think all we need to do is put the calls that send a message to the symbol server into an `@async` block and communicate finish status via a channel back.

In any case, I think we should not do any of this for the extension version 0.13. Right now we have so many bugs in other places that I'd rather have us fix those first and ship a more stable version of the extension next :)

ZacLN:
I understand you feel tied to the current comms system given the work that's been done to get it working but I still feel strongly that we shouldn't be re-implementing an async/parallel framework ourselves. 

Do you have an example of a package that loads differently in a child process I could see because I just don't find that to be a strong argument, at the moment. From my perspective it seems to be overly concerned with a corner case.

I agree re 0.13 stability but the SymbolServer comms interface has made its own contribution to this...

davidanthoff:
I think whatever we decide about this, we should defer it to the v0.14 version of the extension. I think we need to ship v0.13.0 soon to support Julia 1.3, and this seems to big of a change to slip in at this point.

> I agree re 0.13 stability but the SymbolServer comms interface has made its own contribution to this

I think it has been bug free since about mid September with the v0.12.3 release of the extension, at least I'm not aware of any bug related to it right now. I think the crashes we are seeing right now are all related to other problems.

ZacLN:
OK lets discuss after 13.0

PR #65, opened 2019-12-17
Fix a segfault bug

davidanthoff:
Without that we seem to get a fair number of segfaults of the symbol server... Not sure what the root cause is, but this at list fixes it for now.

### DIFF
--- a/src/symbols.jl
+++ b/src/symbols.jl
@@ -99,7 +99,13 @@ function read_methods(x)
         if path === nothing
             path = ""
         end
-        args = Base.invokelatest(Base.arg_decl_parts, m)[2][2:end]
+        args = try
+            Base.invokelatest(Base.arg_decl_parts, m)[2][2:end]
+        catch err
+            # This is a hack around some problem where Julia segfaults
+            # in the line above
+            Tuple{String,String}[]
+        end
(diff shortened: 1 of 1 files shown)


tkf:
Why don't you do this at the "outer most" level (just around IPC boundary and/or logging?). See how REPL and CoreLogging do this:

https://github.com/JuliaLang/julia/blob/a0b67dc0c08afe7ccb268057cc87b0d5405f4a63/stdlib/REPL/src/REPL.jl#L173-L186

https://github.com/JuliaLang/julia/blob/a0b67dc0c08afe7ccb268057cc87b0d5405f4a63/base/logging.jl#L342-L350

The evil part of `show(::IO, ::Type{<:MyType})` overloading is that it recursively kicks in during error handling whenever you try to print a stacktrace.  I don't think fixing the "root cause" is the right direction (because you can't; it was in Setfield in this case).  I think focusing on giving up reporting stacktrace would be a better direction.

davidanthoff:
So we generally try to keep our `try catch` blocks as small as possible, otherwise we might also catch errors in our own code and then we don't get error reports for it anymore. I'll take a look at these two examples, maybe we can still improve the story here, thanks for posting!

PR #69, opened 2020-01-02
copy base/stdlib doc view methods to prevent overloading

ZacLN:
Copies methods to display documentation, etc from /base and REPL/src/docview.jl to prevent overloading by packages (e.g. Flux defined a show method for an internal DataType that caused crashes).


davidanthoff:
Hm, that is a lot of code now that we have to keep in sync with base, right? Maybe we should just heckle these packages to not do such invasive things when this shows up in crash reports? And I think the old code prevents us from crashing, right?

ZacLN:
I think this should be merged, following our discussion the other day?

PR #70, opened 2020-01-02
Don't repeatedly store methods from other modules

ZacLN:
Where functions are overloaded by a package this PR only stores methods associated with the package (before we would duplicate method stores for all methods - when things like `+` are extended this is significantly costly).

Note: Would require changes to StaticLint/LanguageServer logic

### DIFF
--- a/src/symbols.jl
+++ b/src/symbols.jl
@@ -126,7 +126,7 @@ function read_methods(x, M)
     ms = methods(x)
     ms1 = MethodStore[]
     for m in ms
-        !_parentmodules_comp(m.module, M) && parentmodule(x) != m && continue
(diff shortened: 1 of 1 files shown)


davidanthoff:
I think we should merge this soon, and then tag v3.0.0 and start to fix the integration on the LS side, right? I'm waiting with tagging v3.0.0, though, because this PR here seems breaking, right?

ZacLN:
This requires more of a think (it would slow down reference resolution significantly in StaticLint). Not sure it's breaking really but I've modified the PR so that there's no change to package caches but it still makes the internal changes (so you can tag a release).

davidanthoff:
Ok, so I should tag now, without merging this, right?

ZacLN:
No, lets merge it

PR #72, opened 2020-01-12
Rework API for async use

davidanthoff:
@ZacLN this is where I am right now.

Still needs new tests, and some more methods in `utils.jl` to make it work on all Julia versions and crash reporting integration.

Fixes https://github.com/julia-vscode/SymbolServer.jl/issues/73. It now uses the tree hash, SAH of the path name or uuid for the cache filename (for normal packages, deved packages and stdlib packages respectively).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,17 @@ Documentation for working with Julia environments is available [here](https://gi
 ## API
 
 ```julia
-SymbolServerProcess(path_to_env)
+SymbolServerInstance(path_to_env)
 ```
 Launches a server process (with given enviroment) and retrieves the active context. This client side process (this) contains a depot of retrieved packages.
 
 ```julia
-change_env(ssp::SymbolServerProcess, env_path::String)
+change_env(ssp::SymbolServerInstance, env_path::String)
 ```
 Activates a new environment on the server. The new active context must then be retrieved separately.
 
 ```julia
-get_context(ssp::SymbolServerProcess)
+get_context(ssp::SymbolServerInstance)
 ```
 Retrieves the active context (environment) from the server.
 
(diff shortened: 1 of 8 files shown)


ZacLN:
I've had a look and tried it - all seems good. For testing i merged the `baseshow` branch as caching packages was taking forever (and had to shift the `_lookup` funcs to `symbols.jl`). I think the LanguageServer side of this needs a small change to run a scopepass/linting on all files following a reset of the env in the "julia/activateEnvironment" request.

One thing I'm unclear of is whether we can get info on progress from the symserver (i.e. so we can use that progress indicator in vscode.

Otherwise, I'm of a mind to merge this set of PRs as soon as!

davidanthoff:
Great!

> merged the baseshow branch

Yep, lets merge that right after this one here is in, right?

> had to shift the _lookup funcs to symbols.jl

I fixed that here.

>  I think the LanguageServer side of this needs a small change to run a scopepass/linting on all files following a reset of the env in the "julia/activateEnvironment" request.

Probably also easiest done in a follow up PR?

> One thing I'm unclear of is whether we can get info on progress from the symserver (i.e. so we can use that progress indicator in vscode.

Not at the moment, but I have a fairly simple idea how we could add that in a follow up to this design here. I guess one question though is whether we can actually show anything like percent? Maybe we can even just start out with some busy indicator, without specific progress numbers? That we should be able to do with the current design from the LS side of things (we could just send the start notification when we call `getstore`, and then end whenever the results channel has something, right?).

> Otherwise, I'm of a mind to merge this set of PRs as soon as!

Yep! I'm stuck right now getting the tests and code work for older Julia versions... I think there is some logic in the big VERSION branch in `utils.jl` that I should be able to hook into, but somehow I'm not figuring it out... Any chance you could take a look at that?

ZacLN:
Yep I'll have a look

PR #77, opened 2020-01-16
Async citests

ZacLN:
Testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,17 @@ Documentation for working with Julia environments is available [here](https://gi
 ## API
 
 ```julia
-SymbolServerProcess(path_to_env)
+SymbolServerInstance(path_to_env)
 ```
 Launches a server process (with given enviroment) and retrieves the active context. This client side process (this) contains a depot of retrieved packages.
 
 ```julia
-change_env(ssp::SymbolServerProcess, env_path::String)
+change_env(ssp::SymbolServerInstance, env_path::String)
 ```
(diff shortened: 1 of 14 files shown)


ZacLN:
@davidanthoff, this should fix the fiddly compat issues with v1.0- but tests seem to hang indefinitely at the `take!` call in the tests, not sure

davidanthoff:
I added code that shows exceptions in the async task (instead of just having them disappear in the nothing), I think that should help us pin down what is going on. I think there is something that we expect a UUID but get a string...

davidanthoff:
I fixed two more bugs on Julia 1.0, but now have to run. But there should be proper error output at this point that makes it fairly straightforward to find the right fix.

ZacLN:
Tests should pass but Base64 doesn't seem to get pulled in on v1.0- ?

davidanthoff:
Well, we're getting close :) Something wrong on Julia 1.1 only now?

ZacLN:
It's fixed (locally, tests running now) but mostly be eliding the issue, its not clear to me how important the cache naming.

davidanthoff:
I think (hope) that https://github.com/julia-vscode/SymbolServer.jl/pull/77/commits/18946b1b671691c2725b2eee24bf8403a2b782e7 will fix this properly on all versions of Julia. Let's see what CI says :)

ZacLN:
Well done, that was pretty painful getting that to work!

davidanthoff:
DONE! I'll merge, ok?

PR #91, opened 2020-02-10
remove is_stdlib refs

ZacLN:
This no longer seems to be used?


ZacLN:
Hmm, we don't seem to be testing against master or 1.4.0-rc1, can we?

PR #111, opened 2020-03-09
Don't crash when a UUID is missing in manifest

davidanthoff:
Fixes a problem seen in crash reporting.

Also removes the return value, as far as I can tell it is not used anywhere.


ZacLN:
Ignore that last comment - it get's automatically converted back to a string

Issue #108, opened 2020-03-01
Could we index in the cloud?


davidanthoff:
Here is another idea for our indexing woes: could we create the cache files in the cloud for every known tagged package version in the registry, and then the extension would just download these precomputed files? Maybe this could even be integrated with the new package server stuff that is coming in some way.

This would require that the file format doesn’t use serialization so that it becomes Julia version independent. But that would in general be useful if we were to pursue the compiled LS idea.

Also, there seem two use cases:
- user opens a file that is part of a package (say via goto definition) and we want to provide some level of code navigation, hover etc in that file. I think we should try to use LSIF for this scenario in some way.
- user has a `using Foo` Statement in a file, and we need symbols, docs etc from Foo to provide features in the currently open file. I don’t think LSIF helps in that scenario, so we would probably need our own symbol file format for that.

In my ideal world these files maybe would even be just downloaded with the package itself via the package manager, so that the our extension can just rely on them being there. But that is probably something for a future iteration.

non-Jedi:
Could you clarify the "LSIF" acronym please?

davidanthoff:
https://microsoft.github.io/language-server-protocol/specifications/lsif/0.5.0/specification/. I have to admit I don't understand what MS is up-to with that...

PR #123, opened 2020-03-30
More fully describe datatypes

ZacLN:
This PR picks up a fair bit of information that is not currently stored:
* A mirrored type-system is created. This allows proper representation of types through out the symbol-stores. For example, we now have full type information for method arguments.
* No need for the duplication of `show` methods (this was causing crashes where faulty show methods for package datatypes had been defined).
* Return types for methods - would allow us to do more interesting inference stuff.

On the down side:
* Getting return types for methods requires the compilation of everything - time consuming. Caching in the cloud (#108) will probably be necessary to make this acceptable for the avg user (calculation of return types could be disabled in the short term.)

@davidanthoff, this is more for discussion than immediate merge as changes would  across the languagesever/statliclint setup


davidanthoff:
Cool! This also generally means that we are using our own types to represent all of this, which might make it easier to switch the cache file format to say some JSON thing, right? If we do want to go in the cloud direction, then using a file format that is not serialization is probably something we'll have to do... Would also help with the idea to compile the whole LS.

I don't understand the return type part for methods :) How would we decide for which input types we compute return types?

ZacLN:
This is ready now (+ SS/LS relevant PRs).
On the return types - this is disabled for now (too time consuming). Use of return types would require a bit on inference our side (i.e. `sin(1.0)::Float64` as a really simple example)

src/SymbolServer.jl
```diff
@@ -89,7 +90,7 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
 
     take!(server_is_ready)
 
-    p = open(pipeline(Cmd(`$jl_cmd --code-coverage=$(use_code_coverage==0 ? "none" : "user") --startup-file=no --compiled-modules=no --history-file=no --project=$environment_path $server_script $(ssi.store_path) $pipename`, env = env_to_use), stderr = stderr_for_client_process), read = true, write = true)
+    p = open(pipeline(Cmd(`$jl_cmd --code-coverage=$(use_code_coverage==0 ? "none" : "user") --startup-file=no --history-file=no --project=$environment_path $server_script $(ssi.store_path) $pipename`, env = env_to_use), stderr = stderr_for_client_process), read = true, write = true)
```
davidanthoff, line 93: Why remove the no compile mode here? I think there are two benefits to it: 1) indexing packages that aren't precompiled yet is faster, 2) we won't risk precompile race conditions (if two processes try to precompile the same package, Julia errors currently).

ZacLN: Ah, I disabled this for deving, will reverse!

ZacLN:
Yep, sounds good

PR #130, opened 2020-04-16
turn off deprecation warnings for server

ZacLN:


davidanthoff:
This doesn't help. I actually don't think that these messages originate from the symserver process, I think they are created in `load_core()`, and because there is the global const definition that runs during precompile.

ZacLN:
Ah right, that would make sense. Are these warnings an issue? There's no way to avoid triggering them (we want to cache deprecated symbols) but is it OK to turn off dep warnings for the main vscode process?

davidanthoff:
For the CI runs it is a bit annoying because it is a lot of output that has no information (I only pasted a short version, it is more like page after page). I don't think we could easily turn off deprecation warnings for the CI test runs, and we probably don't want to, given that we would to see legit deprecation warnings, right?

All these warnings also appear in the output channel in the VS Code UI, which is maybe the bigger issue. I think there we can just start the LS process without deprecation warnings, though, and that should solve it.

I think medium term we probably have to move the indexing of the stdlib and base out of the LS process and into the symbol server process as well, if we want to statically compile the LS process to make it start faster, and I assume at that point this problem would be automatically solved. Obviously that is a bigger project.

So maybe for now we just start the LS process in VS Code without dep warnings?

ZacLN:
> So maybe for now we just start the LS process in VS Code without dep warnings?

Yep that sounds right.

PR #138, opened 2020-04-24
fix FakeTypeName comparisons

ZacLN:


src/faketypes.jl
```diff
@@ -110,7 +110,7 @@ isfakebottom(t::FakeTypeofBottom) = true
 function Base.:(==)(a::FakeTypeName, b::FakeTypeName)
     a.name == b.name && length(a.parameters) == length(b.parameters) || return false
     for i = 1:length(a.parameters)
-        a.parameters[i] === b.parameters[i] && return false
+        a.parameters[i] == b.parameters[i] || return false
     end
     return true
```
davidanthoff, line 115: Couldn't this whole thing just be written as `a.name == b.name && a.parameters == b.parameters`?

ZacLN: Does comparison of vectors short circuit?

davidanthoff: I think so.

PR #145, opened 2020-04-30
make sure we cache `parse`

ZacLN:


ZacLN:
You've made me question this... and it's wrong. I thought there was something specific about parse being available in each module but not listed in the names but the issue much more generic. Other PRs to follow..

PR #164, opened 2020-06-08
propogate julia version to worker process

ZacLN:
I'm not clear if this is all that needs to be done.

Also to note - SymbolServer does not work with Julia 1.6.

### DIFF
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -24,7 +24,8 @@ end
 function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, progress_callback = nothing, error_handler = nothing)
     !ispath(environment_path) && return :success, recursive_copy(stdlibs)
 
-    jl_cmd = joinpath(Sys.BINDIR, Base.julia_exename())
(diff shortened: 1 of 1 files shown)


src/SymbolServer.jl
```diff
@@ -24,7 +24,8 @@ end
 function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, progress_callback = nothing, error_handler = nothing)
     !ispath(environment_path) && return :success, recursive_copy(stdlibs)
 
-    jl_cmd = joinpath(Sys.BINDIR, Base.julia_exename())
+    # jl_cmd = joinpath(Sys.BINDIR, Base.julia_exename())
+    jl_cmd = Base.julia_cmd().exec[1] # Make sure we use the same Julia version in the worker process
```
pfitzseb, line 28:
```suggestion
    jl_cmd = first(Base.julia_cmd()) # Make sure we use the same Julia version in the worker process
```

I'm not sure how `joinpath(Sys.BINDIR, Base.julia_exename())` can ever be different from `first(Base.julia_cmd())` though...

PR #166, opened 2020-06-08
show method locations

aviatesk:
make it more consistent with Base's print:
```julia
Base.bitshow is a Function.

  # 2 methods for generic function "bitshow":
  [1] bitshow(io::IO, B::BitArray) in Base at show.jl:2256
  [2] bitshow(B::BitArray) in Base at show.jl:2265
```

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -316,6 +316,7 @@ function Base.print(io::IO, m::MethodStore)
         i != length(m.sig) && print(io, ", ")
     end
     print(io, ")")
+    print(io, " in ", m.mod, " at ", normpath(m.file), ':', m.line)
 end
 
 function Base.print(io::IO, t::DataTypeStore)


pfitzseb:
We should be able to convert those paths into clickable links, no? This is interpreted as markdown by VSCode after all.

aviatesk:
no, I tried that at first, but LSP doesn't support `MarkdownString`, which can make links to [invoke vscode commands](https://code.visualstudio.com/api/references/vscode-api#MarkdownString):

It actually only supports `MarkedString` and `MarkupContent` currently: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_hover

pfitzseb:
[Argh...](https://code.visualstudio.com/api/references/vscode-api#MarkedString)

So a normal markdown link like `[show.jl:2256](~/blah/show.jl:2256)` doesn't work?

aviatesk:
ops, vscode seems to automatically handles those links (i.e. open editors in vscode).
Will try to improve.

aviatesk:
But making those links should be done in LangaugeServer.jl side, so I think this PR can be merged as is.

PR #172, opened 2020-06-17
attempt to disable revise if loaded

ZacLN:
https://github.com/julia-vscode/julia-vscode/issues/1343#issuecomment-645482240

### DIFF
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -28,7 +28,7 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
     server_script = joinpath(@__DIR__, "server.jl")
 
     env_to_use = copy(ENV)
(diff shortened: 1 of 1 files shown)


src/SymbolServer.jl
```diff
@@ -28,6 +28,7 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
     server_script = joinpath(@__DIR__, "server.jl")
 
     env_to_use = copy(ENV)
+    ENV["JULIA_REVISE"] = "manual" # Try to make sure Revise isn't enabled.
```
davidanthoff, line 31:
```suggestion
    env_to_use["JULIA_REVISE"] = "manual" # Try to make sure Revise isn't enabled.
```
We want this set for the symbol server process, right? Then it needs to go into `env_to_use`.

PR #175, opened 2020-06-24
Performance improvements

ZacLN:
This rips off most of @timholy's caching approach, from an earlier PR but ensures we pull in all the variables needed (including, inter alia, some imported and using'ed variables). It also fixes a bug whereby within a module a lot of symbols were being stored when they didn't need to be.

Speedup is ~3x for caching some arbitrary packages on my machine.

### DIFF
--- a/src/symbols.jl
+++ b/src/symbols.jl
@@ -53,14 +53,14 @@ struct DataTypeStore <: SymStore
     exported::Bool
 end
 
-function DataTypeStore(t::DataType, parent_mod, exported)
+function DataTypeStore(@nospecialize(t::DataType), parent_mod, exported)
     parameters = map(t.parameters) do p
         _parameter(p)
     end
     types = map(t.types) do p
         FakeTypeName(p)
     end
-    DataTypeStore(FakeTypeName(t), FakeTypeName(t.super), parameters, types, t.isconcretetype && fieldcount(t) > 0 ? collect(fieldnames(t)) : Symbol[], cache_methods(t, parent_mod, exported), _doc(t), exported)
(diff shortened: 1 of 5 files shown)


src/symbols.jl
```diff
@@ -94,64 +98,80 @@ function clean_method_path(m::Method)
     return normpath(path)
 end
 
-function cache_methods(f, mod = nothing, exported = false)
+const _global_method_cache = IdDict{Any,Vector{Any}}()
+function methodinfo(@nospecialize(f); types=Tuple, world=typemax(UInt))
```
timholy, line 102: I added this but I think it's possible to do even better. The key observation is that if you go through all functions just once, then you shouldn't need this. As you loop over names in modules, fetch the object and if it's a Callable you can check whether you've processed it before: just keep a running `IdSet` of all previously-processed functions and check `f ∈ processed`. If not, `push!(procssed, f)`, process the whole MethodTable in one go (assigning methods to different modules) and then move on to the next name; if it is already in `processed`, skip to the next name. That will ensure you call `methods(f)` only once for each `f`.

davidanthoff:
@ZacLN this needs a merge conflict resolution.

src/symbols.jl
ZacLN: This is largely what is being done. Caching of methods should only happen (with a few exceptions) when a function/datatype is encountered from within its parent module. [This](https://github.com/julia-vscode/SymbolServer.jl/pull/175/files#diff-6a3b27e5bfe405fa75cb4196464e0adcR160-R174) then distributes those methods across the different `FunctionStore`s

PR #178, opened 2020-06-29
add :Main to stdlibs

ZacLN:
:Main is now referenced within the cache Base so needs to be present


davidanthoff:
Hm, tests crash with a segfault...

PR #185, opened 2020-08-06
improved name resolution

pfitzseb:
The newly added `all_names` has a few advantages over `split_module_names`:
1. Seems to be about an order (or two) of magnitude faster:
```
julia> @time all_names(JSON);
  0.000248 seconds (51 allocations: 7.281 KiB)

julia> @time SymbolServer.split_module_names(JSON, allns);
  0.015986 seconds (19.93 k allocations: 1.255 MiB)
```
2. It seems to be more correct (in this specific case its `Base.parse` being shadowed by `JSON.Parsers.parse`, which is imported into `JSON` with `using .Parsers: parse`):
```
julia> :parse in all_names(JSON)
true

julia> a, b = SymbolServer.split_module_names(JSON, allns);

julia> :parse in a || :parse in b
false
```

This is only a very superficial integration of that function, mostly because I'm not 100% confident I haven't missed anything and would like to hear @ZacLN's opinion on this approach first :)

@timholy might also be interested in this, given his recent work on performance.

### DIFF
--- a/src/symbols.jl
+++ b/src/symbols.jl
@@ -330,17 +330,18 @@ function getenvtree(names = nothing)
 end
 
 # faster and more correct split_module_names
-function all_names(m, pred = x -> true, symbols = Set(Symbol[]), seen = Set(Module[]))
+all_names(m) = all_names(m, x -> isdefined(m, x))
+function all_names(m, pred, symbols = Set(Symbol[]), seen = Set(Module[]))
     push!(seen, m)
-    ns = unsorted_names(m; all = true, imported = true)
+    ns = unsorted_names(m; all = true, imported = false)
     for n in ns
         isdefined(m, n) || continue
         Base.isdeprecated(m, n) && continue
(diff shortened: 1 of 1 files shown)


ZacLN (approved):

ZacLN:
Side note - could we all try to keep formatting changes to separate PRs?

pfitzseb:
Yeah, sorry about that -- seems like VSCode is autoformatting files on save :)

aviatesk:
We turn on auto-formatting on save [here](https://github.com/julia-vscode/julia-vscode/blob/69d4b1a539baae28ec4f21e1b7659d2a8485bc42/.vscode/settings.json#L18), but we may want to make that TS-specific.

pfitzseb:
Eh, I like that behaviour. It's just that the kwarg option is different between what I have locally set and this codebase. So either we fix that in `.vscode` or reformat our Julia code.

PR #190, opened 2020-10-10
feat: get_symbol_server

aminya:
Adds a function to get symbols for the given module or current environment

### DIFF
--- a/.github/workflows/jlpkgbutler-ci-pr-workflow.yml
+++ b/.github/workflows/jlpkgbutler-ci-pr-workflow.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
+    fail-fast: false
     strategy:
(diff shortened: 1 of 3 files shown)


aminya:
@davidanthoff @ZacLN I have added the tests for this. Although all the tests pass online, it fails in the CI! Could you check this?

davidanthoff:
What is the use-case for this? I'm a bit hesitant to add more public facing APIs in general, and in this case I don't understand what this would be used for :)

aminya:
> What is the use-case for this? I'm a bit hesitant to add more public facing APIs in general, and in this case I don't understand what this would be used for :)

See https://github.com/julia-vscode/StaticLint.jl/pull/220#discussion_r552232228

davidanthoff:
Do we still need this PR, or is the plan with a new external Lint.jl package superseding this?

PR #191, opened 2020-10-22
Check if submod.vals has key before retrieving

quinnj:
This seems to fix https://github.com/julia-vscode/LanguageServer.jl/issues/840 for me locally


aviatesk (approved):
lgtm (not sure why this has never been reported for those packages with lazily computed fields though.)

pfitzseb (approved):

ZacLN (approved):

PR #194, opened 2020-12-17
fix on Julia nightly

simeonschaub:
see https://github.com/JuliaLang/julia/pull/38835 and https://github.com/JuliaLang/julia/pull/38136

### DIFF
--- a/.github/workflows/jlpkgbutler-ci-master-workflow.yml
+++ b/.github/workflows/jlpkgbutler-ci-master-workflow.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5']
+        julia-version: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5', 'nightly']
         julia-arch: [x64, x86]
         os: [ubuntu-latest, windows-latest, macOS-latest]
         exclude:
--- a/src/faketypes.jl
+++ b/src/faketypes.jl
@@ -15,6 +15,9 @@ struct FakeTypeName
 end
 
 function FakeTypeName(@nospecialize(x); justname=false)
+    @static if !(Vararg isa Type)
+        x isa Core.TypeofVararg && return FakeTypeofVararg(x)
+    end
     if x isa DataType
         xname = x.name
         xnamename = xname.name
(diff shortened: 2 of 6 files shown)


pfitzseb (dismissed):
LGTM, but @ZacLN should probably take a look at this too.

simeonschaub:
Hmm, ok, I am now not quite sure how `FakeTypeName` is supposed to work. Is it expected that something like `Array{Int}` just drops the parameters for the `DataType` `Array{Int,TypeName(:N)}` completely? It seems to be intentional because `FakeTypeName` is called with `justname=true`.

ZacLN (approved):

simeonschaub:
Bump. What should be done about the tests here?

src/faketypes.jl
```diff
@@ -120,3 +123,47 @@ Base.:(==)(a::FakeTypeVar, b::FakeTypeVar) = a.lb == b.lb && a.name == b.name &&
...
+        N
+        FakeTypeofVararg() = new()
+        FakeTypeofVararg(T) = new(T)
+        FakeTypeofVararg(T, N) = new(T, N)
+    end
+    function FakeTypeofVararg(va::Core.TypeofVararg)
+        if isdefined(va, :N)
+            FakeTypeofVararg(FakeTypeName(va.T), va.N)
+        elseif isdefined(va, :T)
+            FakeTypeofVararg(FakeTypeName(va.T))
```
simeonschaub, line 139: Is this more correct?
```suggestion
            FakeTypeofVararg(FakeTypeName(va.T; justname=true), va.N)
        elseif isdefined(va, :T)
            FakeTypeofVararg(FakeTypeName(va.T; justname=true))
```

musm:
bump, would love to be able to use this on 1.7

ZacLN:
> Hmm, ok, I am now not quite sure how `FakeTypeName` is supposed to work. Is it expected that something like `Array{Int}` just drops the parameters for the `DataType` `Array{Int,TypeName(:N)}` completely? It seems to be intentional because `FakeTypeName` is called with `justname=true`.

Yes, in it's current state this isn't intended to fully describe the types. Parameters are dropped except where at the first level of a DataType. For example given `abstract type T{S} end` we would have a full representaiton of `T{Bool}` but not of `T{T{Bool}}` and so `FakeTypeName(T{T{SomeType}}) == FakeTypeName(T{T{SomeOtherType}})` is expected. This is to prevent storing 'massive' datatypes - there are some huge ones in LinearAlgebra. Given this, I don't think the tests are quite appropriate (i.e. they're asking too much) and happy for them to be dropped.

I've not had a look at the new approach to Varargs and a different heuristic may well be appropriate.

simeonschaub:
I still kept most of the tests, but excluded some in earlier Julia versions. I hope they are not too strict now, but my thought was that they can always be loosened if they become a problem. This also now passes on `justname=true` for all type parameters of `Vararg` to be consistent with the previous behavior.

> This is to prevent storing 'massive' datatypes - there are some huge ones in LinearAlgebra.

Do you mean things like `StridedArray`? Maybe we could piggyback off the type alias printing added to Base in https://github.com/JuliaLang/julia/pull/36107 here, so we only store the alias if there is a matching one.

ZacLN (approved):

ZacLN:
Thanks, looks good. Yes exactly that, I'll have a look at the PR (thanks for the ref).

simeonschaub:
Good to merge?

musm:
Would also be great to have a tag with the fix.

ZacLN:
Yes, staticlint will need small adjustments.

.github/workflows/jlpkgbutler-ci-master-workflow.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5']
+        julia-version: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5', 'nightly']
```
davidanthoff, line 13: We should generally not edit these files manually, they are managed by package butler and any change we make manually here will be undone by package butler down the road.

simeonschaub: Oops, sorry! Didn't realize they were automatically generated.

davidanthoff: No harm done, it will just disappear at some point again :)

PR #202, opened 2021-01-13
Add registry indexer

davidanthoff:
Very WIP. Some things that still need to be done:
- [x] Only index and save one package per call to the main script
- [x] Fall back to older Julia versions if there is an error (deferred)
- [x] Record failed trials so that we can skip things
- [x] Skip things that are already cached
- [x] Come up with a file structure for the cached results
- [x] Figure out how to handle stdlib and base (deferred)
- [x] Don't write separate cache versions for different Julia versions
- [x] Compress things

### DIFF
--- a/src/indexpackage.jl
+++ b/src/indexpackage.jl
@@ -1,13 +1,21 @@
 using Pkg, SHA
 using Base: UUID
 
-current_package_name = ARGS[1]
+current_package_name = Symbol(ARGS[1])
 current_package_version = VersionNumber(ARGS[2])
-current_package_treehash = ARGS[3]
+current_package_uuid = UUID(ARGS[3])
+current_package_treehash = ARGS[4]
+
+@info "Indexing package $current_package_name $current_package_version..."
+
+current_package_versionwithoutplus = replace(string(current_package_version), '+'=>'_')
+cache_package_folder_path = joinpath(store_path, "v1", "packages", "$(current_package_name)_$current_package_uuid")
+cache_path = joinpath(cache_package_folder_path, "v$(current_package_versionwithoutplus)_$current_package_treehash.jstore")
+
+mkpath(cache_package_folder_path)
 
 module LoadingBay end
 Pkg.add(name=current_package_name, version=current_package_version)
-
 
 # TODO Make the code below ONLY write a cache file for the package we just added here.
 include("faketypes.jl")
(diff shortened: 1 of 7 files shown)


src/indexpackage.jl
```diff
@@ -0,0 +1,148 @@
...
+end
+
+symbols(env_symbols, nothing, getallns(), visited)
+
+# Wrap the `ModuleStore`s as `Package`s.
+for (pkg_name, cache) in env_symbols
+    pkg_name = String(pkg_name)
+    !isinmanifest(ctx, pkg_name) && continue
+    uuid = packageuuid(ctx, String(pkg_name))
+    pe = frommanifest(ctx, uuid)
```
ZacLN, line 141:
```julia
if !haskey(Base.loaded_modules, Base.PkgId(uuid, pkg_name))
    @info "can't save $pkg_name"
    continue
end
modify_dirs(cache, f -> modify_dir(f, pkg_src_dir(Base.loaded_modules[Base.PkgId(uuid, pkg_name)]), "PLACEHOLDER"))
```

We need to strip the package paths ^

ZacLN:
I'm not sure of the benefits of changing how we store the caches for Base/Core at the moment outweigh the costs (i.e. within the SymbolServer module, cached as part of the compilation process). Loading from disc would be slower. 

I can see the possibility in the future of wanting to use SymbolServer loaded in a vXX Julia process as part of an environment representing Julia vYY, but we're not there yet. 

We also have the option of caching the stdlibs alongside Base/Core in the SymbolServer module.

src/indexpackage.jl
```diff
@@ -0,0 +1,73 @@
...
+    LoadingBay.eval(:(import $current_package_name))
+    getfield(LoadingBay, current_package_name)
+catch e
+    @info "Could not load package, exiting."
+    exit(10)
+end
+
+# Get the symbols
+env = getenvtree([current_package_name])
+symbols(env, m)
```
ZacLN, line 52: when https://github.com/julia-vscode/SymbolServer.jl/pull/203 is merged to master (and the commit pulled across to here) this should be changed to `symbols(env, m, get_return_type = true)`

davidanthoff:
> I'm not sure of the benefits of changing how we store the caches for Base/Core at the moment outweigh the costs (i.e. within the SymbolServer module, cached as part of the compilation process). Loading from disc would be slower. Arguments for doing this would include running type inference on all methods which is almost certianly too time consuming to be the default option for most users.

Ah, great, so we can skip that for now!

> I can see the possibility in the future of wanting to use SymbolServer loaded in a vXX Julia process as part of an environment representing Julia vYY, but we're not there yet.

Agreed.

> We also have the option of caching the stdlibs alongside Base/Core in the SymbolServer module.

The downside of that is probably that during precompile we would now precompile all the stdlibs, right? But maybe the best strategy for now to just get all of this out, right?

davidanthoff:
I'm going to merge this as is, given that it can't break anything. And then we can iterate with more improvements in future PRs.

PR #203, opened 2021-01-13
propogate get_return_types

ZacLN:
We will want to turn this on once we have remote indexing up and running.


davidanthoff:
Just out of curiosity: how does this actually work? If we want to propagate return types, don't we need to know for which input types? Are we just using the same logic that precompile is using?

ZacLN:
This is just turning on the `get_return_types` option properly - making sure it gets passed to everything as we cache. Nothing fancy actually relating to the inference

pfitzseb (approved):

PR #205, opened 2021-01-15
Refer to Core.TypeofVararg as typeof(Vararg)

mkitti:
See https://github.com/JuliaLang/julia/pull/39239#issuecomment-760016914

`Core.TypeofVararg` introduced in https://github.com/JuliaLang/julia/pull/3813
`VERSION >= v"1.7.0-DEV.77"`


pfitzseb (approved):
Thanks!

PR #212, opened 2021-02-01
add Intrinsics methods

ZacLN:


src/symbols.jl
```diff
@@ -533,6 +533,13 @@ function load_core(; get_return_type = false)
     haskey(cache[:Core], :_typebody) && push!(cache[:Core][:_typebody!].methods, MethodStore(:_typebody!, :Core, "built-in", 0, [:a => FakeTypeName(Any), :b => FakeTypeName(Any)], Symbol[], FakeTypeName(Any)))
     push!(cache[:Core][:(===)].methods, MethodStore(:(===), :Core, "built-in", 0, [:a => FakeTypeName(Any), :b => FakeTypeName(Any)], Symbol[], FakeTypeName(Any)))
     push!(cache[:Core][:(<:)].methods, MethodStore(:(<:), :Core, "built-in", 0, [:a => FakeTypeName(Type{T} where T), :b => FakeTypeName(Type{T} where T)], Symbol[], FakeTypeName(Any)))
+    # Add unspecified methods for Intrinsics, working out the actual methods will need to be done by hand?
```
pfitzseb, line 536: Could parse https://github.com/JuliaLang/julia/blob/master/src/intrinsics.h for the number of arguments, but that seems more trouble than it's worth.

pfitzseb (approved):

PR #213, opened 2021-02-08
store methods for shadowed datatypes

ZacLN:
For next minor - stores methods for datatypes that have been imported into a module and rebound under a new name. 

For next minor version rather than patch.

Fixes one of the issues in https://github.com/julia-vscode/julia-vscode/issues/1224


pfitzseb (approved):

davidanthoff:
Does this mean we have to trigger reindexing on client machines?

ZacLN:
Ideally we'd want all caches updated

PR #214, opened 2021-02-16
create dummy cache locally for downloads of unsuccessful caches

ZacLN:


pfitzseb:
Doesn't this disable local indexing? I feel like we could just write the `.unavailable` file into the cache dir and then check for its existence before downloading, right?

ZacLN:
That makes more sense

pfitzseb (approved):

PR #219, opened 2021-04-13
remove limit on type names

ZacLN:
Store full names of DataTypes. This was disable previously as it slows caching but this should be reversed if we're moving to the cloud approach.

### DIFF
--- a/src/faketypes.jl
+++ b/src/faketypes.jl
@@ -14,23 +14,18 @@ struct FakeTypeName
     parameters::Vector{Any}
 end
 
-function FakeTypeName(@nospecialize(x); justname=false)
+function FakeTypeName(@nospecialize(x))
     @static if !(Vararg isa Type)
         x isa typeof(Vararg) && return FakeTypeofVararg(x)
     end
     if x isa DataType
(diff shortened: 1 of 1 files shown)


davidanthoff:
Needs a merge conflict resolution.

Does this affect our cache store? If so, we need to cycle this in with a cache store version update, right? So v1.2 would be an excellent vehicle for that because we are updating the cache version for that already.

Also, if this does change things that are stored in the cache, we need to update our cloud computed/cached files as well with that? I.e. reindex everything?

ZacLN:
This does both (so needs new local cache store version + recaching on cloud). It adds a computational cost so my intention had been to merge it at the same time as we disable local caching - is that v1.2?

davidanthoff:
No, that won't be 1.2, so I removed it from that milestone for now.

davidanthoff:
@ZacLN Now that we are doing cloud indexing, should we do this?

ZacLN:
Yep let's do it

davidanthoff:
@ZacLN Alright, just needs a merge resolve!

PR #220, opened 2021-04-13
turn on get_return_types in indexpackage.jl

ZacLN:
Runs inference on all methods when we're caching them (costly)


davidanthoff:
@ZacLN should we do this now? But do we need to reindex everything then?

ZacLN:
Let's turn it on. Not necessary to re-cachr everything as no down stream code would be dependent on it.

Caching will become quite a bit more expensive.

PR #227, opened 2021-07-16
Allow passing the project file as environment.

fredrikekre:


pfitzseb (approved):
Thanks!
Test failure is unrelated and should be fixed by https://github.com/julia-vscode/SymbolServer.jl/pull/228.

davidanthoff:
Why do we want to do that? I'm generally not a fan of making APIs behave a la "you can pass it either as a folder or file path, we'll just handle both cases", that seems to just add unnecessary complexity. And in this case we've used the directory path forever in the API, and the VS Code client won't pass anything else. Could other clients not just conform with that as well?

I do like the other aspects of the PR, certainly is a bit more elegant :)

fredrikekre:
It's just nice to be able to pass the output of `Base.active_project()` I guess?

It also seems that the full path is "more canonical"  since otherwise the library has to guess what the user meant (although having both Project.toml and JuliaProject.toml is pretty uncommon).

davidanthoff:
I think if this is the canonical way, then we should change everything to make this the only way of specifying environments.

davidanthoff (changes requested):
So I think if someone wants to change this throughout the LS and SymbolServer, then we could make that change, so that we just accept file names everywhere. It seems a fairly involved change, though, and at the moment I think we have such a backlog in terms of bugs that we probably should try to fix those first...

fredrikekre:
> accept file names everywhere

This is the only place where it is *not* accepted to pass the file I believe.

PR #232, opened 2021-07-25
Handle boo return from env_project_file

davidanthoff:
Hopefully fixes a bug from crash reporting. We generally need to handle `Bool` return values from `env_project_file`, not sure what I did here is the right behavior, though...

@ZacLN can you review, I think you originally wrote this code.


pfitzseb (approved):
This looks correct. We might want to add support for implicit envs (`true` return` at some point though.

ZacLN (approved):

davidanthoff:
I don't even know what an implicit env is :)

PR #245, opened 2021-11-17
Fix missing docs for some symbols

BenPH:
This adds documentation for UnionAlls, Union types, and Modules. Also, aliases will have documentation that is associated with them and not from of whatever they are aliasing.

Fixes https://github.com/julia-vscode/julia-vscode/issues/2564.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -179,11 +179,18 @@ function sha_pkg(pe::PackageEntry)
     path(pe) isa String && isdir(path(pe)) && isdir(joinpath(path(pe), "src")) ? sha2_256_dir(joinpath(path(pe), "src")) : nothing
 end
 
-_doc(object::UnionAll) = _doc(Base.unwrap_unionall(object))
 function _doc(@nospecialize(object))
+    local binding
     try
         binding = Base.Docs.aliasof(object, typeof(object))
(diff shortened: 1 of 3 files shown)


src/utils.jl
```diff
@@ -179,10 +179,8 @@ function sha_pkg(pe::PackageEntry)
     path(pe) isa String && isdir(path(pe)) && isdir(joinpath(path(pe), "src")) ? sha2_256_dir(joinpath(path(pe), "src")) : nothing
 end
 
-function _doc(@nospecialize(object))
+function _doc(binding::Base.Docs.Binding)
```
pfitzseb, line 182: Any specific reason for this change? The `@nospecialize` may have been important(-ish) for performance.

pfitzseb (approved):
LGTM. Thanks!

src/utils.jl
BenPH: No reason – I'm just not familiar with `@nospecialize`. The function does expect a binding now. Should it be `@nospecialize(binding)` or `@nospecialize(binding::Base.Docs.Binding)`?

pfitzseb: The former. But this is fine as-is, since it's only ever be called with a `Binding` anyways.

PR #246, opened 2021-12-01
Try to keep pipenames within bounds on unix.

fredrikekre:


src/SymbolServer.jl
```diff
@@ -156,6 +156,27 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
...
+        return "\\\\.\\pipe\\vscjlsymserv-$(UUIDs.uuid4())"
+    end
+    # Pipe names on unix may only be 92 chars (JuliaLang/julia#43281), and since
+    # tempdir can be arbitrary long (in particular on macos) we try to keep the name
+    # within bounds here.
+    n = "vscjlsymserv-$(string(UUIDs.uuid4())[1:8])"
+    pipename = joinpath(tempdir(), n)
+    if length(pipename) >= 92
+        # Try to use /tmp and if that fails, hope the long pipename works anyway
+        tmp = "/tmp/vscjlsymserv"
```
pfitzseb, line 170: Should probably have *some* random component in here so multiple processes don't conflict (and the path can definitely be longer than this).

fredrikekre: This is just a directory, I add 8 chars of uuid below

pfitzseb: Ah, true.

fredrikekre: It is not really necessary except for a way to check that the user have permissions, but not sure how to check that otherwise.

fredrikekre: Updated it to `touch(f); rm(f)` since that won't leave any files around.

fredrikekre:
Okay, tests should pass now: apparently Julia < 1.3 doesn't respect `TEMPDIR`.

pfitzseb (approved):

PR #254, opened 2022-06-27
Set version to 7.3.0.

fredrikekre:


davidanthoff:
Please use the tag functionality in VS Code to bump versions. otherwise we end up in a state that that doesn't allow the next tag to work smoothly. If you don't use VS Code, just ping me and I can do the bump.

PR #263, opened 2023-01-27
Symbol caches: Larger download batching. Don't ask for non-General pkgs

IanButterworth:
Given non-General packages don't have symbol caches, don't request them, to avoid requesting URLs that include the UUID & name of private packages.

Also the download batching didn't make sense to me. It seems better to limit the concurrent number of downloads, rather than the total number of batches. Update: Now batches of 100

Fixes #.

- [ ] End-user documentation check. If this PR requires end-user documentation in the Julia VS Code extension docs, please add that at https://github.com/julia-vscode/docs.
- [ ] Changelog mention. If this PR should be mentioned in the CHANGELOG for the Julia VS Code extension, please open a PR against https://github.com/julia-vscode/julia-vscode/blob/master/CHANGELOG.md with those changes.

### DIFF
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -31,7 +31,7 @@ function get_general_pkgs()
     @static if VERSION >= v"1.7-"
         regs = Pkg.Types.Context().registries
         i = findfirst(r -> r.name == "General" && r.uuid == UUID("23338594-aafe-5451-b93e-139f81909106"), regs)
-        i === nothing && error("Could not find the General registry")
+        i === nothing && return Dict{UUID, PkgEntry}()
         return regs[i].pkgs
     else
         for r in Pkg.Types.collect_registries()
@@ -39,7 +39,7 @@ function get_general_pkgs()
             reg = Pkg.Types.read_registry(joinpath(r.path, "Registry.toml"))
             return reg["packages"]
         end
-        error("Could not find the General registry")
+        return Dict{UUID, PkgEntry}()
     end
 end
 
@@ -48,9 +48,17 @@ end
 
 Removes packages that aren't going to be on the symbol cache server because they aren't in the General registry.
 This avoids leaking private package name & uuid pairs via the url requests to the symbol server.
+
+If the General registry cannot be found packages cannot be checked, so all packages will be removed.
 """
 function remove_non_general_pkgs!(pkgs)
     general_pkgs = get_general_pkgs()
+    if isempty(general_pkgs)
+        @warn """
+        Could not find the General registry when checking for whether packages are public.
+        All package symbol caches will be generated locally"""
+        return empty!(pkgs)
+    end
     filter!(pkgs) do pkg
         packageuuid(pkg) === nothing && return false
         packagename(pkg) === nothing && return false
@@ -88,6 +96,7 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
                         try
                             remove_non_general_pkgs!(to_download)
                         catch err
+                            # if any errors, err on the side of caution and mark all as private, and continue
                             @error """
                             Symbol cache downloading: Failed to identify which packages to omit based on the General registry.
                             All packages will be processsed locally""" err
@@ -95,15 +104,18 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
                         end
                         if !isempty(to_download)
                             n_done = 0
-                            @sync for pkg in to_download
-                                @async begin
-                                    yield()
-                                    uuid = packageuuid(pkg)
-                                    get_file_from_cloud(manifest, uuid, environment_path, ssi.depot_path, ssi.store_path, download_dir, ssi.symbolcache_upstream)
-                                    yield()
-                                    n_done += 1
-                                    percentage = round(Int, 100*(n_done)/length(to_download))
-                                    progress_callback !== nothing && progress_callback("Downloading caches...", percentage)
+                            n_total = length(to_download)
+                            @sync for batch in Iterators.partition(to_download, 100) # 100 connections at a time
+                                for pkg in batch
+                                    @async begin
+                                        yield()
+                                        uuid = packageuuid(pkg)
+                                        get_file_from_cloud(manifest, uuid, environment_path, ssi.depot_path, ssi.store_path, download_dir, ssi.symbolcache_upstream)
+                                        yield()
+                                        n_done += 1
+                                        percentage = round(Int, 100*(n_done/n_total))
+                                        progress_callback !== nothing && progress_callback("Downloading caches...", percentage)
+                                    end
                                 end
                             end
                             progress_callback !== nothing && progress_callback("All cache files downloaded.", 100)
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -105,8 +105,8 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
                         if !isempty(to_download)
                             n_done = 0
                             n_total = length(to_download)
-                            @sync for batch in Iterators.partition(to_download, 100) # 100 connections at a time
-                                for pkg in batch
+                            for batch in Iterators.partition(to_download, 100) # 100 connections at a time
+                                @sync for pkg in batch
                                     @async begin
                                         yield()
                                         uuid = packageuuid(pkg)
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -28,18 +28,26 @@ mutable struct SymbolServerInstance
 end
 
 function get_general_pkgs()
-    @static if VERSION >= v"1.7-"
-        regs = Pkg.Types.Context().registries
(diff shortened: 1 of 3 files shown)


src/SymbolServer.jl
```diff
@@ -27,6 +27,17 @@ mutable struct SymbolServerInstance
     end
 end
 
+function remove_non_general_pkgs!(pkgs)
+    reg_pkgs = Pkg.Types.Context().registries[3].pkgs
```
davidanthoff, line 31: I think this needs to be changed to find wich element in `registries` is actually the general registry, right? On my system for example the length of it is just 1, so I assume you have two additional registries on your system?

src/SymbolServer.jl
```diff
@@ -27,6 +27,17 @@ mutable struct SymbolServerInstance
     end
 end
 
+function remove_non_general_pkgs!(pkgs)
```
davidanthoff, line 30: Could you add a test for this function? Any code that interacts with the internals of `Pkg` often needs to be adapted for different Julia versions, and for that it is really helpful to have a test that then runs during CI on all the different Julia versions we need to support.

src/SymbolServer.jl
```diff
@@ -42,7 +53,8 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
                     if manifest !== nothing
                         @debug "Downloading cache files for manifest at $(manifest_filename)."
                         to_download = collect(validate_disc_store(ssi.store_path, manifest))
-                        batches = Iterators.partition(to_download, max(1, floor(Int, length(to_download)÷50)))
+                        remove_non_general_pkgs!(to_download)
+                        batches = Iterators.partition(to_download, 10) # download up to 10 at a time
```
davidanthoff, line 57: This definitely seems better than the previous version, but I'm actually starting to wonder why we batch at all? Maybe we should just fire all requests off at the same time async? My guess is that at most this is something like a few hundred HTTP requests, which doesn't strike me as a lot?

davidanthoff (commented):
Cool, thanks! Left a few comments.

src/SymbolServer.jl
fredrikekre, line 30: Can make this a more general filter function that also filter out packages with `nothing` as UUID, no point in trying  those URLs either.

src/SymbolServer.jl
IanButterworth: Whoops, yes, I meant to leave a comment asking how you'd typically get General in the extension

src/SymbolServer.jl
IanButterworth: Ok, updated to no batching. I guess it's worth trying on insider

src/SymbolServer.jl
IanButterworth: I added this, but AFAICT because pkgs comes from the manifest.deps, which is a `Dict{Base.UUID, Pkg.Types.PackageEntry}()` it's not possible for the UUID to be nothing. But there is support here for <1.1 julia, which I struggle to follow (and question why it's still supported..) so I'll leave these checks in

fredrikekre: I just know I have seen URLs with `nothing` in the LS.jl logs, even on recent Julias. Perhaps for stdlibs?

IanButterworth: I think I have too, but I can't figure out how a pkg without uuid could end up in `pkgs` here. @davidanthoff any idea? It'd be good to test it

IanButterworth: Seems like it was the `tree_hash` being nothing that we were seeing, i.e. stdlibs, so I added filtering and testss

src/SymbolServer.jl
```diff
@@ -27,6 +27,35 @@ mutable struct SymbolServerInstance
...
+        All packages will be processed locally.
+        """
+        empty!(pkgs)
+        return pkgs
+    end
+    reg = Pkg.Types.Context().registries[i]
+    filter!(pkgs) do pkg
+        packageuuid(pkg) === nothing && return false
+        packagename(pkg) === nothing && return false
+        tree_hash(pkg) === nothing && return false # stdlibs don't have tree_hash and aren't cached
```
fredrikekre, line 50: Should also apply to devved packages, but maybe that is filtered out earlier.

IanButterworth:
This has proven quite painful to get working on older julia versions. I went as far as 1.1, and for 1.0 rely on a try-catch to default to just generated caches locally. (Perhaps someone else can see a fix for 1.0 ?)

IanButterworth:
@davidanthoff it would be great, if you approve, to get this trialed on the pre-release

src/SymbolServer.jl
```diff
@@ -27,6 +27,49 @@ mutable struct SymbolServerInstance
     end
 end
 
+function get_general_pkgs()
+    @static if VERSION >= v"1.7-"
+        regs = Pkg.Types.Context().registries
+        i = findfirst(r -> r.name == "General" && r.uuid == UUID("23338594-aafe-5451-b93e-139f81909106"), regs)
+        i === nothing && error("Could not find the General registry")
```
fredrikekre, line 34: Instead of error here and try-catch further down I suggest returning an empty package list from this function. It isn't necessarily something wrong if you don't have General installed.

src/SymbolServer.jl
```diff
@@ -42,20 +85,29 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
...
+                            remove_non_general_pkgs!(to_download)
+                        catch err
+                            @error """
+                            Symbol cache downloading: Failed to identify which packages to omit based on the General registry.
+                            All packages will be processsed locally""" err
+                            empty!(to_download)
+                        end
+                        if !isempty(to_download)
+                            n_done = 0
+                            @sync for pkg in to_download
```
fredrikekre, line 98: Should the number of concurrent download tasks be limited? Perhaps doesn't matter.

fredrikekre: Yea maybe not needed.

ericphanson: Adding a limit makes sense to me; I think you can get stuff like requests timing out if there’s tons and the cpu can’t service them all

src/SymbolServer.jl
IanButterworth: Makes sense. Updated

src/SymbolServer.jl
IanButterworth: Ok. Updated to batches of 100

src/SymbolServer.jl
```diff
@@ -42,20 +93,33 @@ function getstore(ssi::SymbolServerInstance, environment_path::AbstractString, p
...
+                            @error """
+                            Symbol cache downloading: Failed to identify which packages to omit based on the General registry.
+                            All packages will be processsed locally""" err
+                            empty!(to_download)
+                        end
+                        if !isempty(to_download)
+                            n_done = 0
+                            n_total = length(to_download)
+                            @sync for batch in Iterators.partition(to_download, 100) # 100 connections at a time
+                                for pkg in batch
```
ericphanson, line 109: I think the sync has to go inside, so we do a whole batch of 100 before moving to the next, right?
```suggestion
                            for batch in Iterators.partition(to_download, 100) # 100 connections at a time
                                @sync for pkg in batch
```

src/SymbolServer.jl
```diff
@@ -27,6 +27,57 @@ mutable struct SymbolServerInstance
     end
 end
 
+function get_general_pkgs()
+    @static if VERSION >= v"1.7-"
+        regs = Pkg.Types.Context().registries
+        i = findfirst(r -> r.name == "General" && r.uuid == UUID("23338594-aafe-5451-b93e-139f81909106"), regs)
+        i === nothing && return Dict{UUID, PkgEntry}()
+        return regs[i].pkgs
```
IanButterworth, line 35: This isn't working in the vscode extension because `Pkg.Types.Context().registries` is empty
```
┌ Info: regs
└   regs = Pkg.Registry.RegistryInstance[]
┌ Error: Symbol cache downloading: Failed to identify which packages to omit based on the General registry.
│ All packages will be processsed locally
│   err = Could not find the General registry
└ @ SymbolServer ~/Documents/GitHub/julia-vscode/scripts/packages/SymbolServer/src/
```

IanButterworth: What's the right way to get the General registry from here?

pfitzseb: That's kinda tricky. We're overwriting setting the `JULIA_DEPOT_PATH` env variable to a depot vendored in the extension, but you could probably temporarily change the DEPOT_PATH before this?

IanButterworth: Ok. I've tried to implement that. I'd appreciate trying to get this merged to stop sending out non-General package name & uuids

IanButterworth:
Will do, thanks

IanButterworth:
Bump 🙏

pfitzseb (approved):

pfitzseb:
Thanks for the PR! I'll make a few follow-up tweaks, but none of that should hold this up any further :)

IanButterworth:
Just checking, is this used on the extension now? I wasn't sure if the extension uses a non-tagged commit

Issue #269, opened 2023-05-22
Extreme inefficiency in main server loop


thomasjm:
I was looking at this code in `server.jl`, where it loads the packages one by one, and then *every time* does a bunch of work over the whole package environment:

https://github.com/julia-vscode/SymbolServer.jl/blob/3162b6a1b53d7036c53e24c3cc2fe21b90acf67a/src/server.jl#L95-L100

The comment alludes to how this may duplicate some work. It is a bit of an understatement. I was testing with indexing an environment containing only `Plots` and its dependencies, and I tried moving the `end` of the loop on line `95` up to line `97`, so we do all the `load_package` calls up front.

This made the total indexing time go from **109.1 seconds** to **12.2 seconds**. Almost a 10X improvement from this one change!

In general, I think the perf of `SymbolServer.jl` has more low-hanging fruit, such as indexing and loading on-disk stores in parallel using multiple threads. I feel that the current painful performance has led to workarounds like the cloud indexing system, which I've never seen before in any other LSP ecosystem and which I really wish didn't exist. (Even more philosophically, I wonder why gobs of symbols need to be indexed and stored in memory up-front at all? Doesn't Julia have the introspection machinery to enable `LanguageServer.jl` to look up (and cache as needed) the necessary symbols as it runs?)

In my usage at least, `SymbolServer.jl` is the main contributor to slow startup times and high CPU/memory usage when using `LanguageServer.jl`. I think a leaner and meaner `SymbolServer.jl` would make life much nicer for `LanguageServer.jl` users.

CC @mkitti since you were so helpful with the last perf issue I reported. Also CC @davidanthoff.

mkitti:
Could you send a pull request to test?

thomasjm:
This may be a bit more than you asked for, but I've been meaning to open a PR with a collection of improvements I've been working on: #270. With this PR, it's more straightforward to just call `SymbolServer.index_packages(...)` in order to test the performance, since that logic isn't hidden in a script.

If you just want the change that this issue is about, it looks like this: https://github.com/julia-vscode/SymbolServer.jl/compare/master...codedownio:SymbolServer.jl:indexing-perf?expand=1

thomasjm:
Oh and if you use Nix, I can provide a Nix-based way to test the indexing.

PR #270, opened 2023-05-22
Make SymbolServer.jl more library-like

thomasjm:
This is an effort to make `SymbolServer.jl` more library-like, in the sense that significant logic is not hidden in script files like `indexpackage.jl` and `server.jl`. This logic is now exposed in two exported functions: `SymbolServer.index_package` and `SymbolServer.index_packages`, to make it easier for third party packages to use.

There are also some miscellaneous spelling and grammar fixes, as well as a few `@time` calls to help with profiling the indexing process. It's probably easiest to review this commit-by-commit.

- [X] End-user documentation check. If this PR requires end-user documentation in the Julia VS Code extension docs, please add that at https://github.com/julia-vscode/docs.
- [X] Changelog mention. If this PR should be mentioned in the CHANGELOG for the Julia VS Code extension, please open a PR against https://github.com/julia-vscode/julia-vscode/blob/master/CHANGELOG.md with those changes.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,144 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.8.5"
+manifest_format = "2.0"
+project_hash = "e9b501a7a2bfe6bb16747a24c09b28861fd51dd6"
+
+[[deps.ArgTools]]
+uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+version = "1.1.1"
+
+[[deps.Artifacts]]
+uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
+
+[[deps.Base64]]
+uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+
+[[deps.Dates]]
+deps = ["Printf"]
+uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
+
+[[deps.Downloads]]
+deps = ["ArgTools", "FileWatching", "LibCURL", "NetworkOptions"]
+uuid = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
+version = "1.6.0"
+
+[[deps.FileWatching]]
+uuid = "7b1f6079-737a-58dc-b8bc-7a2ca5c1b5ee"
+
+[[deps.InteractiveUtils]]
+deps = ["Markdown"]
+uuid = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
+
+[[deps.LibCURL]]
+deps = ["LibCURL_jll", "MozillaCACerts_jll"]
+uuid = "b27032c2-a3e7-50c8-80cd-2d36dbcbfd21"
+version = "0.6.3"
+
+[[deps.LibCURL_jll]]
+deps = ["Artifacts", "LibSSH2_jll", "Libdl", "MbedTLS_jll", "Zlib_jll", "nghttp2_jll"]
+uuid = "deac9b47-8bc7-5906-a0fe-35ac56dc84c0"
+version = "7.84.0+0"
+
+[[deps.LibGit2]]
+deps = ["Base64", "NetworkOptions", "Printf", "SHA"]
+uuid = "76f85450-5226-5b5a-8eaa-529ad045b433"
+
+[[deps.LibSSH2_jll]]
+deps = ["Artifacts", "Libdl", "MbedTLS_jll"]
+uuid = "29816b5a-b9ab-546f-933c-edad1886dfa8"
+version = "1.10.2+0"
+
+[[deps.Libdl]]
+uuid = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+
+[[deps.Logging]]
+uuid = "56ddb016-857b-54e1-b83d-db4d58db5568"
+
+[[deps.Markdown]]
+deps = ["Base64"]
+uuid = "d6f4376e-aef5-505a-96c1-9c027394607a"
+
+[[deps.MbedTLS_jll]]
+deps = ["Artifacts", "Libdl"]
+uuid = "c8ffd9c3-330d-5841-b78e-0817d7145fa1"
+version = "2.28.0+0"
+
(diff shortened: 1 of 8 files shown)


mkitti:
What is with the disc -> disk thing? Isn't this merely British versus American spelling?

thomasjm:
No, I put a citation in the commit message:

> Disk is indisputably the correct spelling for computer storage drives

https://grammarist.com/spelling/disc-disk/

This is also the case in Britain; the article provides a link to the Telegraph with an example.

thomasjm:
I wouldn't normally open a whole PR just to be pedantic about spelling :P. I'm happy to split those commits off if you prefer. They would produce a merge conflict with the current changes though.

> So, moving all this code into the regular package will essentially make it unavailable to both the local indexing process as started by LS and the cloud indexing service that runs inside the docker container.

Hmm, I'll have to think about this a bit more. Why is it important that `SymbolServer` not be available in the indexing process? I notice that `server.jl` currently imports almost all of the same code anyway. If it could be made available, users of `SymbolServer` wouldn't have to do weird stuff like [this](https://github.com/julia-vscode/SymbolServer.jl/blob/3162b6a1b53d7036c53e24c3cc2fe21b90acf67a/src/SymbolServer.jl#L66).

Maybe we could just move the indexing logic into separate `.jl` files, which can then be `include`d by both the main `SymbolServer` module and the application scripts?

On the general idea of launching things in separate processes: I don't love it, because due to Julia's precompilation, every additional process tends to lead to significant additional latency and memory usage. Also, the fact the `SymbolServer` has to communicate with the main `LanguageServer` over a socket using an ad-hoc protocol adds complexity and a bit of a barrier to parallelizing the indexing. I understand the desire to re-nice the server process as done [here](https://github.com/julia-vscode/SymbolServer.jl/blob/3162b6a1b53d7036c53e24c3cc2fe21b90acf67a/src/server.jl#L11-L23), but I'm not actually sure the separate process is a net win, especially if `SymbolServer` could be optimized generally (e.g. see #269).

davidanthoff:
> Why is it important that `SymbolServer` not be available in the indexing process?

We want to run the indexing in exactly the same environment that the user has set up. And that environment almost certainly won't have `SymbolServer` in it, so we can't use standard package loading mechanisms to access it.

> Maybe we could just move the indexing logic into separate `.jl` files, which can then be `include`d by both the main `SymbolServer` module and the application scripts?

Yes, I think that could work, but really the design of the package `SymbolServer.jl` is that it is _not_ loaded inside the process that is doing the actual indexing. The design is that `SymbolServer.jl` is the package that runs in a separate process that a) triggers indexing (in a child process) and b) gives access to the cached symbol information.

> On the general idea of launching things in separate processes

The general idea here is this: we _never_ want to load or run any user-code inside the language server process. Otherwise, the LS would crash every time there is a bug in a package that a user loads. That is not an option for us, the LS provides too many capabilities that work even if a user tries to load a package with an error to completely give up in these situations. The primary reason for the two-process model is to isolate and harden the language server process against errors in packages that users use.

mkitti:
> I wouldn't normally open a whole PR just to be pedantic about spelling :P. I'm happy to split those commits off if you prefer. They would produce a merge conflict with the current changes though.

PRs with just spelling or grammar changes are much easier to merge than ones mixed in with technical changes. There should not be a merge conflict with the changes are exactly the same.

thomasjm:
Thanks for explaining @davidanthoff! It seems this PR isn't mergeable for now so I'll close it. I opened #271 with the spelling stuff.

> The general idea here is this: we never want to load or run any user-code inside the language server process. Otherwise, the LS would crash every time there is a bug in a package that a user loads.

Possibly naive question, but aren't there tricks to work around this? I'm no Julia expert but I came across [this method](https://stackoverflow.com/a/36327924) for loading a package using metaprogramming. (It presumably could be combined with a try/catch block, and even `eval`-ed in an isolated `LoadingBay`-style module of your choice.)

davidanthoff:
> Possibly naive question, but aren't there tricks to work around this?

I don't think so. Most basic situation: what if user code runs `exit()`? But in general Julia really has zero mechanisms to run a specific piece of code in some kind of sandbox within a given Julia process, so a separate process really is the only option. We certainly could catch _some_ errors, but I don't see any way to make it robust against all errors (which the current solution is).

thomasjm:
Hmm, a library calling `exit()` at import-time seems pretty out there. If the vast majority of errors could be caught by a try/catch block, maybe there's a still tradeoff to be made here.

However--has anyone ever thought about doing this analysis statically, without loading code at all? I was looking at [JuliaSyntax.jl](https://github.com/JuliaLang/JuliaSyntax.jl) and it seems like a pretty impressive project, and advertises that one of its goals (besides speed) is to support tooling. 

Static analysis is how most LSP implementations work after all. As a point of comparison: Python is similar to Julia in many ways, as an interpreted language which has the ability to run side-effectful code at import time. All the Python LSP servers I'm aware of rely on static analysis using a library such as [jedi](https://github.com/davidhalter/jedi). The now-deprecated [Microsoft Python language server](https://github.com/microsoft/python-language-server) wasn't even written in Python. Could there be a future where Julia's language server works the same way?

davidanthoff:
> Hmm, a library calling exit() at import-time seems pretty out there. If the vast majority of errors could be caught by a try/catch block, maybe there's a still tradeoff to be made here.

Yeah, that is probably not the best example. But for example, any binary shared library that any package loads that crashes would take down the entire language server. And when that crashes, _a lot_ of things in the extension stop working, so that is just not an option.

> However--has anyone ever thought about doing this analysis statically, without loading code at all?

Yes, at the moment we are doing a hybrid: we look at the current project, and for anything that is deved we use static analysis, and for all other packages we extract all information via the SymbolServer.jl (either from a cloud cache or by loading it locally once and then caching locally). The problem with static analysis is that Julia is a _very_ dynamic language and we essentially stand no chance of really ever fully getting it right. Things like `@eval` just evade any attempt at static analysis. In my mind, the current hybrid story is the right compromise: for things that don't change quickly (i.e. at each key press) we extract _all_ the information that there is by loading the package at some stage, and for things that change very quickly (i.e. files are edited in real time) we use static analysis, that won't get everything right, but is fast.

PR #272, opened 2023-09-01
Bugfix: methods added to functions in other packages missing in cache

brychcy:
This reverts part of commit 50521fb.

Fixes [vscode#2565](https://github.com/julia-vscode/julia-vscode/issues/2565) and the problems related to DataFrames reported in comments to [vscode#1576](https://github.com/julia-vscode/julia-vscode/issues/1576)

### DIFF
--- a/src/server.jl
+++ b/src/server.jl
@@ -89,44 +89,39 @@ for (pk_name, uuid) in toplevel_pkgs
     end
 end
 
-visited = Base.IdSet{Module}([Base, Core])
-
 # Load all packages together
+# This is important, or methods added to functions in other packages that are loaded earlier would not be in the cache
 for (i, uuid) in enumerate(packages_to_load)
     load_package(ctx, uuid, conn, LoadingBay, round(Int, 100*(i - 1)/length(packages_to_load)))
-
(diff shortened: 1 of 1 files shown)


brychcy:
Note: After fixing this, I had to delete  ~/Library/'Application Support'/Code/User/globalStorage/julialang.language-julia/ (begin o macOS) and restart. Using ">Julia: Re-Index Language Server Cache" was not enough.

brychcy:
Example code for reproduction: Setup a fresh project, just add "DataFrames". Add a file with the following code:
```
using DataFrames

x = DataFrame(a=[1, 2])
y = nrow(x)
print(y)
```

Remove the cache directory mentioned in the previous comment and restart vscode.

A "Possible method call error" error is reported  for "nrow". Cmd-Clicking on "nrow" doesn't show a definition.

Now repeat with the patch applied. The errors disappears and, cmd-clicking shows the definitions in DataFrames.

pfitzseb (approved):
Thanks!

It's a bit unfortunate that we write cache files in one go now, which means that one package failing to load means that you'll never get any info. I'll merge this and do some thinking on how to improve that situation later.

Issue #285, opened 2024-02-17
Symbol cache staleness when using a `develop`ed package


tpgillam:
This issue describes behaviour observed in neovim and LanguageServer.jl, via julia-lsp. Having dug around in the code a bit, it looks like the caching logic is here rather than `LanguageServer.jl`, hence filing here.

## Case 1 (the issue)

Consider the following file structure within some directory.

```
├── A
│   ├── Project.toml
│   └── src
│       └── A.jl
└── proj
    ├── Manifest.toml
    ├── Project.toml
    └── test.jl
```

There is a Julia package called "A", and beside it is a project called "proj".
`A` is a package generated with `import Pkg; Pkg.generate("A")` from the top-level directory.

As-of Julia 1.10.1, there is a single `A.greet()` method defined.

Within proj we run `] dev ../A`, such that its Manifest.toml contains:
```toml
[[deps.A]]
path = "../A"
uuid = "94f385dd-073b-49fe-b7ed-f824d09b3331"
version = "0.1.0"
```

Within `proj/test.jl` we write:
```julia
using A

A.greet()
```

And the language server can find `A` and `A.greet`, so all is well.

Now, close add a new function into `A/src/A.jl`, e.g. 
```julia
moo() = 42
```

Back in `proj/test.jl`, now extend the file to call the new function.
The language server now declares the new symbol as missing:
```julia
using A

A.greet()
A.moo()    # Missing reference: moo
```

Restarting the language server (or the whole editor) doesn't change anything.

We can see that when the language server starts, it loads `A` from the cache, but doesn't detect that the cache is stale:
```
"[ Info: Starting LS at 1708163760\n"
"[ Info: Package A (94f385dd-073b-49fe-b7ed-f824d09b3331) is cached.\n"
"[ Info: Symbol server indexing took 5.283324494 seconds.\n"
"[ Info: Loading A from cache... (0%)\n"
"[ Info: Done loading A from cache... (took 0.12s) (0%)\n"
"[ Info: Loaded all packages into cache in 0.12s (100%)\n"
```

## Case 2 (similar, but works)

Now we consider a slightly different case, where `A` is now _underneath_ the project directory, as follows:
```
├── A
│   ├── Project.toml
│   └── src
│       └── A.jl
├── Manifest.toml
├── Project.toml
└── test.jl
```

In this situation the issue above is not seen. Instead, if we edit `A/src.jl` and then open `test.jl`, when the language server starts we see:
```
[...]

PR #286, opened 2024-02-17
Correctly compute project SHA when Manifest contains paths starting with `../`

tpgillam:
Fixes https://github.com/julia-vscode/SymbolServer.jl/issues/285

This PR ensures that the project SHA is correct computed when an environment's Manifest contains relative paths starting with `../`.

It also adds a test that would have caught the issue above.

To the reviewer: I bumped the patch version, since I'm used to releasing on every PR, but do advise what would be standard. (Naturally we're keen for the issue to be resolved in LanguageServer soon!)

---

- [ ] Changelog mention. If this PR should be mentioned in the CHANGELOG for the Julia VS Code extension, please open a PR against https://github.com/julia-vscode/julia-vscode/blob/master/CHANGELOG.md with those changes.


pfitzseb (dismissed):
LGTM

pfitzseb:
Let's disable nightly CI temporarily so we can get this merged (unless you wanna debug and fix that failure too).

tpgillam:
> Let's disable nightly CI temporarily so we can get this merged (unless you wanna debug and fix that failure too).

Ha, I think I'll pass on looking at Julia nightly builds -- but why don't I add a commit just to temporarily disable that one? Then we can convince ourselves that nothing is broken at least.

pfitzseb (approved):

.github/workflows/jlpkgbutler-ci-pr-workflow.yml
```diff
@@ -9,7 +9,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['1.0', '1.6', '1.9', 'nightly']
+        # FIXME: temporarily removing nightly due to failing build.
```
tpgillam, line 12: Should we revert this before merging, now that the tests have passed?

pfitzseb: Yeah, let's.

tpgillam: Done.

PR #296, opened 2025-02-25
Fix UndefVarError for PkgEntry

### LINKED ISSUE #294: Errors if JULIA_DEPOT_PATH is set
Sorry if I'm missing something but the logic here seems like only `~/.julia` will be checked, regardless of what `Base.DEPOT_PATH` or `JULIA_DEPOT_PATH` is?:

https://github.com/julia-vscode/SymbolServer.jl/blob/2201c26a43013a7bf71b4829ee4967e072070cab/src/SymbolServer.jl#L42

E.g.:

```
julia> ENV["JULIA_DEPOT_PATH"]
"/usr/local/share/julia"

julia> Base.DEPOT_PATH
1-element Vector{String}:
 "/usr/local/share/julia"

julia> SymbolServer.get_general_pkgs()
ERROR: UndefVarError: `PkgEntry` not defined
Stacktrace:
 [1] get_general_pkgs()
   @ SymbolServer ~/.vscode-server/extensions/julialang.language-julia-1.127.2/scripts/packages/SymbolServer/src/SymbolServer.jl:46
 [2] top-level scope
   @ REPL[11]:1
```

The error is triggered because it hits the `i===nothing` branch and then `PkgEntry` wasn't defined. But I think the original problem is the `DEPOT_PATH` logic?

visr:
Fixes #294.

The language server showed this log:

```
┌ Error: Symbol cache downloading: Failed to identify which packages to omit based on the General registry.
│ All packages will be processsed locally
│   err =
│    UndefVarError: `PkgEntry` not defined in `SymbolServer`
│    Suggestion: check for spelling errors or missing imports.
└ @ SymbolServer 
```

This UndefVarError lead to always erroring on the side of caution as mentioned here:

https://github.com/julia-vscode/SymbolServer.jl/blob/2201c26a43013a7bf71b4829ee4967e072070cab/src/SymbolServer.jl#L114-L122

### DIFF
--- a/src/SymbolServer.jl
+++ b/src/SymbolServer.jl
@@ -3,6 +3,7 @@ module SymbolServer
 export SymbolServerInstance, getstore
 
 using Pkg, SHA
+using Pkg.Registry: PkgEntry
 using Base: UUID, Process
 import Sockets, UUIDs
 


pfitzseb (approved):

pfitzseb:
Thanks!

visr:
Would be great to have this merged.

src/SymbolServer.jl
```diff
@@ -51,7 +55,7 @@ function get_general_pkgs()
                 reg = Pkg.Types.read_registry(joinpath(r.path, "Registry.toml"))
                 return reg["packages"]
             end
-            return Dict{UUID, PkgEntry}()
+            return Dict{String, Any}()
```
visr, line 58: Now this empty Dict is the same type as `reg["packages"]`. The difference was already handled in `remove_non_general_pkgs!`.

visr:
Ok I've addressed that now.

Running this code locally still results in `get_general_pkgs()` being empty since my depot isn't in my homedir. I guess that means this doesn't fix #294, just a surface issue on top?

See also this discussion: https://github.com/julia-vscode/SymbolServer.jl/pull/263#discussion_r1097282808

How can we better find the General registry?

